PR #3, opened 2017-08-26
Linking with Stan.jl

Ayush-iitkgp:
This PR is the firt step to integrate with Stan.jl. 
Few future improvements include:
1. Allowing users to input the priors using Distributions.jl (the defualt behaviour is that the parameters follow Normal(0,1)).
2. Output more readable output such as the mean and the quantiles of the parameters. At present, it returns the Mamba chains

### DIFF
--- a/src/bayesian_inference.jl
+++ b/src/bayesian_inference.jl
@@ -1,0 +1,52 @@
+export StanModel, bayesian_inference
+
+struct StanModel{R,C}
+  return_code::R
+  chain_results::C
+end
+function bayesian_inference(prob::DEProblem,t,data;alg="integrate_ode_rk45",num_samples=1, num_warmup=1,kwargs...)
+  length_of_y = string(length(prob.u0))
+  length_of_parameter = string(length(prob.f.params))
+
+  const parameter_estimation_model = "
+  functions {
+    real[] sho(real t,real[] u,real[] theta,real[] x_r,int[] x_i) {
+      real du[$length_of_y];  // 2 = length(prob.u0) = length_of_y
+      // placeholder for differential equation here
+      du[1] = theta[1] * u[1] - 1.0*u[1] * u[2]; //string(prob.f.pfuncs[1])
+      du[2] = -3.0*u[1] + 1.0*u[1] * u[2];
+      return du;
+      }
+    }
+  data {
+    real u0[$length_of_y]; // 2 = length(prob.u0)
+    int<lower=1> T;
+    real u[T,$length_of_y]; // 2 = length(prob.u0)
+    real t0;
+    real ts[T];
+  }
+  transformed data {
+    real x_r[0];
+    int x_i[0];
+  }
+  parameters {
+    vector<lower=0>[$length_of_y] sigma;   // 2 = length(prob.u0)
+    real theta[$length_of_parameter];   // // 1=length(prob.f.params)
+  }
+  model{
+    real u_hat[T,$length_of_y]; // 2 = length(prob.u0)
+    sigma ~ inv_gamma(2, 3);
+    // placeholder for priors here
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Looks good. Can it get some unit tests to make sure the pieces are producing the right strings, and that this generates some chains without errors?

Ayush-iitkgp:
@gabrielgellner @goedman please have a look. Your comments are valuable.

src/bayesian_inference.jl
```diff
@@ -27,11 +27,16 @@ function generate_priors(f)
...
 end
 
-function bayesian_inference(prob::DEProblem,t,data;alg=:integrate_ode_rk45,num_samples=1, num_warmup=1,kwargs...)
+function bayesian_inference(prob::DEProblem,t,data;alg=:rk45,num_samples=1, num_warmup=1,kwargs...)
   length_of_y = string(length(prob.u0))
   f = prob.f
   length_of_parameter = string(length(f.params))
-  alg = string(alg)
+  if alg ==:rk45
+  alg = integrate_ode_rk45
```
ChrisRackauckas: don't these need to be strings?

Ayush-iitkgp:
A slight change in API, now the users can pass distribution of the priors using the Normal distribution from Distributions.jl.

ChrisRackauckas:
I don't think the default `Normal(0,1)` makes sense. Priors should always be required.

src/bayesian_inference.jl
```diff
@@ -0,0 +1,88 @@
...
+    real x_r[0];
+    int x_i[0];
+  }
+  parameters {
+    vector<lower=0>[$length_of_y] sigma;
+    real theta[$length_of_parameter];
+  }
+  model{
+    real u_hat[T,$length_of_y];
+    sigma ~ inv_gamma(2, 3);
```
ChrisRackauckas: This should be set when the user chooses the likelihood somehow.

src/bayesian_inference.jl
```diff
@@ -0,0 +1,88 @@
...
+    $priors_string
+    u_hat = $algorithm(sho, u0, t0, ts, theta, x_r, x_i);
+    for (t in 1:T){
+      u[t] ~ normal(u_hat[t], sigma);
```
ChrisRackauckas: We should allow the choice here somehow. Maybe in a future PR.

src/bayesian_inference.jl
Ayush-iitkgp: This is the default nature. Should ask users to input the distribution of sigma as well?

src/bayesian_inference.jl
Ayush-iitkgp: Very much possible :)
Let's get the Normal Distribution working.

src/bayesian_inference.jl
ChrisRackauckas: For now, yes. Later this should be part of the likelihood choice I think.

ChrisRackauckas: For now, keep this I guess. It'll get addressed when likelihoods are addressed. General likelihoods should have different ways of setting hyperparameters.

Ayush-iitkgp:
The tests are passing on my Mac.

goedman:
I’ll continue looking at the results tomorrow, but looking at the diagnostic plots doesn’t look promising. Also the tests too often fail on my machine, also a Mac.

The number of samples in runtests.jl is also rather low.

But I will look at it again tomorrow.

ChrisRackauckas:
Yeah, it might need to be run longer, but that means we might need to have CI just on Linux so that way it doesn't bog down our CI times. I made those changes, and we should check in more detail what's going on. My guess is it has to do with the fact that negative parameters don't make sense, so I know Turing.jl needed the truncated priors to do well.

PR #10, opened 2017-08-28
Pass option to Stan ODE solvers

Ayush-iitkgp:
Addresses #10 
Looks correct to me but Stan is throwing a compilation error.

SYNTAX ERROR, MESSAGE(S) FROM PARSER:

first argument to integrate_ode_rk45 must be the name of a function with signature (real, real[], real[], real[], int[]) : real[]   error in '/Users/ayush/.julia/v0.6/DiffEqParamEstim/tmp/parameter_estimation_model.stan' at line 30, column 97
  -------------------------------------------------
    28:     sigma ~ inv_gamma(2, 3);
    29:     theta[1] ~ normal(1.5, 1.0) ; 
    30:     u_hat = integrate_ode_rk45(sho, u0, t0, ts, theta, x_r, x_i, rel_tol, abs_tol, max_num_steps);
                                                                                                        ^
    31:     for (t in 1:T){
  -------------------------------------------------

PARSER EXPECTED: ")"

make: *** [/Users/ayush/.julia/v0.6/DiffEqParamEstim/tmp/parameter_estimation_model.hpp] Error 253

### DIFF
--- a/src/bayesian_inference.jl
+++ b/src/bayesian_inference.jl
@@ -35,7 +35,7 @@ function generate_priors(f,priors)
   priors_string
 end
 
-function bayesian_inference(prob::DEProblem,t,data,priors = nothing;alg=:rk45,num_samples=1000, num_warmup=1000,reltol=1e-3,abstol=1e-6,maxiter=Int(1e5),kwargs...)
+function bayesian_inference(prob::DEProblem,t,data,priors = nothing;alg=:rk45,num_samples=1000, num_warmup=1000, reltol=1e-3, abstol=1e-6, maxiter=Int(1e5),kwargs...)
   length_of_y = string(length(prob.u0))
   f = prob.f
   length_of_parameter = string(length(f.params))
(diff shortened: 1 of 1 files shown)


Ayush-iitkgp:
@ChrisRackauckas @goedman want to have a look?

Ayush-iitkgp:
The arguments only needed to be passed to the model part.

Ayush-iitkgp:
Tests pass on my Mac.

ChrisRackauckas:
Looks like the test is just far too sensitive. Likely due to the priors.

goedman:
If you want to make these Stan language variables you probably have to define them in a block.

Latest version of your branch passes.

ChrisRackauckas:
Is there any advantage for making them variables instead of just interpolating them into the string?

goedman:
Would there ever be a case where a Stan program would adapt those? Not sure if that would even be valid. Somewhat akin to building a global stiffness matrix vs an element by element approach in FEM.

For now interpolation is a good solution.

ChrisRackauckas:
>Would there ever be a case where a Stan program would adapt those? Not sure if that would even be valid. Somewhat akin to building a global stiffness matrix vs an element by element approach in FEM.

There are times when it could happen. `AutoAbstol` in the DiffEq Callback Library is one case. But even then, it would be internal to the ODE solver, and Boost and Sundials cannot do this. So at least for now this would never come up.

PR #14, opened 2017-12-26
Update model call syntax for 0.6-compatible Turing

xukai92:


xukai92:
The test for Turing passes on my local.
Seems sound Stand tests broken on Travis...

ChrisRackauckas:
@goedman did something change in the Stan syntax recently?

goedman:
Hi @ChrisRackauckas Not that I'm aware, I did make some updates in master for Julia 0.7. I'll have a look at above test.

goedman:
A first look seems to indicate the generated syntax of the Stan program (line break after minus sign?) is not/no longer (?) accepted. I’ll dig deeper tomorrow.

goedman:
On my system (OS X) the stanc (Stan compiler from Stan language to C++) in Stan 2.17 (and maybe Stan 2.16 on Travis as well?) no longer accepts the ___... in the function section. I'll check on Stan discourse if I can get that confirmed. I'll also have a look at the macro used to generate the function, but I'm pretty weak w.r.t. meta programming.

ChrisRackauckas:
Oh, Stan doesn't allow variables with `__` in front? That was added recently to make sure the macro names don't collide with user names. I am very surprised Stan cannot handle that. Can we just put that in as a feature request? Seems simple enough.

goedman:
Section 4.2 of the Stan language manual indeed states that variables must start wit a ASCII letter (a-z or A-Z). I’ll check with Bob but don’t have high hopes that will change (single underscores in variable names can have specific meaning for distributions). Any chance we add a character in front of the 3 _s?

goedman:
Just looked at the definition of macro ```ode_def_nohes``` and see changing this is more involved. The names are always defined in ```ode_def_opts```? Would you be willing to consider a keyword argument like ```symbol_prefix='___'```?

ChrisRackauckas:
Making a generic prefix was too difficult given the call overloading part. I just changed the internal variable prefix to be hard coded as `internal_var___`. That should have no namespace issues and hopefully Stan's parser can handle that...

Is there an issue to follow for it? I would assume there must be some parser overhaul issue for Stan given that this doesn't work, but I can't find one.

ChrisRackauckas:
The newest version of ParameterizedFunctions.jl prefixes with `internal_var___` and the Stan failure still exists. @goedman can you interpret the error message? Where are you checking the log for this?

goedman:
Hmm, on my system I see:

SYNTAX ERROR, MESSAGE(S) FROM PARSER:

  error in '/Users/rob/.julia/v0.6/Stan/Examples/OtherExamples/DiffEqBayes/Lorenz/tmp/parameter_estimation_model.stan' at line 8, column 7
  -------------------------------------------------
     6:     */
     7:       real du[3];
     8:       ___du[1] = 10.0 * (___u[2] - ___u[1]);
              ^
     9:       du[2] = u[1] * (28.0 - u[3]) - u[2];
  -------------------------------------------------

PARSER EXPECTED: "}"

make: *** [/Users/rob/.julia/v0.6/Stan/Examples/OtherExamples/DiffEqBayes/Lorenz/tmp/parameter_estimation_model.hpp] Error 253

This is in the ...build.log file in the tmp directory when using the Lorenz example in Stan.jl. The version in DiffEqBase creates the tmp dir in the current working dir where Julia was started.

I have not contacted Bob Carpenter yet about this issue yet, but this results seems to support my assumption that the Stan compiler (stanc) takes special action when encountering an _.

Wanted to understand the problem a bit better before raising this issue. But from your earlier remark I take it the ```symbol_prefix='___'``` approach won't work?

Occasionally my responses may seem slow, but that's because I'm in Europe (and will be for the next 2 weeks).

Happy New Year!

ChrisRackauckas:
That's the old version of ParameterizedFunctions.jl. Pkg.update to get 2.3.1 which doesn't have any variables starting with `___`.

goedman:
Don't think so (I did run Pkg.update(), see below in the generated stan code). I actually think it is how stanc handles the triple _s. 

```
julia> Pkg.installed("ParameterizedFunctions")
v"2.3.1"
```

and

```
functions {
    real[] sho(real t,real[] u,real[] theta,real[] x_r,int[] x_i) {
    /*
      real du[3];
          internal_var___du[1] = 10.0 * (internal_var___u[2] - internal_var___u[1]);    internal_var___du[2] = internal_var___u[1] * (28.0 - internal_var___u[3]) - internal_var___u[2];    internal_var___du[3] = internal_var___u[1] * internal_var___u[2] - theta[1] * internal_var___u[3];
    */
      real du[3];
      ___du[1] = 10.0 * (___u[2] - ___u[1]);
      du[2] = u[1] * (28.0 - u[3]) - u[2];
      du[3] = u[1] * u[2] - theta[1] * u[3];
      return du;
      }
    }
  data {
    real u0[3];
    int<lower=1> T;
    real u[T,3];
    real t0;
    real ts[T];
    real reltol;
    real abstol;
    int maxiter;
  }
  transformed data {
    real x_r[0];
    int x_i[0];
  }
  parameters {
    vector<lower=0>[3] sigma;
    real theta[1];
  }
  model{
    real u_hat[T,3];
    sigma ~ inv_gamma(2, 3);
    theta[1] ~ normal(2.6666, 1.0) ; 
    u_hat = integrate_ode_rk45(sho, u0, t0, ts, theta, x_r, x_i, reltol, abstol, maxiter);
    for (t in 1:T){
      u[t] ~ normal(u_hat[t], sigma);
      }
  }
```

ChrisRackauckas:
Oh, does `real du[$length_of_y];` need to be changed?

goedman:
I’m trying to pinpoint what is going wrong. The result I briefly posted is suspect, looks like I had leftover updates in the Stan program. I’ll clean up my (Stan.jl) versions of the examples and try again.

ChrisRackauckas:
I think it's just that we hard-coded the name for `du`:

https://github.com/JuliaDiffEq/DiffEqBayes.jl/blob/master/src/stan_inference.jl#L56

I can test that just by opening a PR (my computer that can run Stan is currently busy). I wonder if it can work with the `___du` case if that's changed? If not it's fine to keep the new prefix.

ChrisRackauckas:
I rebased this PR onto the Stan fix.

ChrisRackauckas:
I added a test and it passed, but only when Turing.jl was on master. Since we aren't registered yet I just merged it, though in the future version control and support needs to be followed more closely.

https://github.com/yebai/Turing.jl/issues/416

PR #17, opened 2018-01-06
Implemented stan_string 

Vaibhavdixit02:


src/stan_string.jl
```diff
@@ -0,0 +1,111 @@
...
+	parameters = (params(p)[1])
+	return string("theta[$i] ~ beta_binomial($parameters)",";")
+end
+function stan_string(p::Normal,i)
+	parameters = (params(p)[1],params(p)[2])
+	string("theta[$i] ~ normal$parameters",";")
+end
+function stan_string(p::TDist,i) 
+	parameters = (params(p)[1])
+	return string("theta[$i] ~ normal($parameters,0,1)",";")
```
ChrisRackauckas: Huh, you're sure Stan uses Normal for thus just with a DoF part?

ChrisRackauckas:
The test timed out. That could mean there's something a little off, but I'll re-run it to see.

src/stan_string.jl
Vaibhavdixit02: Actually Distributions.jl implementation has only a DoF part https://juliastats.github.io/Distributions.jl/latest/univariate.html#Distributions.TDist

ChrisRackauckas: I see. But Stan calls it a Normal distribution?

ChrisRackauckas:
Please check locally if it's correct. Travis seems to be failing but it looks like it should be fine? Could just be random failures, in which case we need to make the test easier for Stan (or use a truncated distribution!). But please double check that the Normal distribution string is the same.

PR #19, opened 2018-01-11
Changes in stan_string and options for likelihood and hyperparameters

Vaibhavdixit02:


src/stan_inference.jl
```diff
@@ -25,16 +25,17 @@ function generate_priors(f,priors)
...
+      priors_string = string(priors_string,"theta[$i] ~",stan_string(priors[i]),";")
     end
   end
   priors_string
 end
 
 function stan_inference(prob::DEProblem,t,data,priors = nothing;alg=:rk45,
                             num_samples=1000, num_warmup=1000, reltol=1e-3,
-                            abstol=1e-6, maxiter=Int(1e5),kwargs...)
+                            abstol=1e-6, maxiter=Int(1e5),likelihood=Normal,vars=("StanODEData",InverseGamma(2,3)))
```
ChrisRackauckas: Instead of a string, use a symbol `:StanODEData` so it gets interned.

ChrisRackauckas: Eh actually you're turning it into a string right after, why both.

PR #26, opened 2018-01-18
plot_chains implementation

Vaibhavdixit02:


Vaibhavdixit02:
I think Travis needs to be re-run for this?

Vaibhavdixit02:
@ChrisRackauckas I think you missed this.

ChrisRackauckas:
You should remake this into a Plots.jl recipe so that it matches the rest of DiffEq. See the type recipes here:

http://docs.juliaplots.org/latest/recipes/

src/utils.jl
```diff
@@ -0,0 +1,11 @@
+using Mamba
```
ChrisRackauckas: These aren't needed since it's inherited from the module it's in.

PR #27, opened 2018-01-18
Initial version of Notebook for Stan

Vaibhavdixit02:
@ChrisRackauckas  this is the first draft of sorts will edit and add as per your review. :smile:


ChrisRackauckas:
Don't use plotly because it doesn't save well

Don't save the extra Stan files

Why are the thetas repeated?

Other than that, looks good. Do this PR to DiffEqBenchmarks.jl instead.

PR #29, opened 2018-01-30
Added 4 parameter LotkaVolterra Test and changes in turing_inference

Vaibhavdixit02:


src/turing_inference.jl
```diff
@@ -15,7 +15,7 @@ function turing_inference(prob::DEProblem,alg,t,data,priors = nothing;
                     InverseGamma(2, 3),
                     Turing.VarName(vi, [:bif, :σ], ""),
                     vi)
-    p_tmp = problem_new_parameters(prob, theta); sol_tmp = solve(p_tmp,alg;kwargs...)
+    p_tmp = problem_new_parameters(prob, theta); sol_tmp = solve(p_tmp,alg,saveat=t;kwargs...)
```
ChrisRackauckas: `saveat` is a keyword argument, so it needs to be on the other side of the `;`.

PR #39, opened 2018-04-12
Pull request/1172005b

marcjwilliams1:
oops, forget to add them in git!


src/abc_inference.jl
```diff
@@ -0,0 +1,24 @@
+function createabcfunction(prob, t, distancefunction, alg, kwargs...)
+    function simfunc(params, constants, targetdata)
+        sol = solve(problem_new_parameters(prob, params), alg, kwargs...)
```
ChrisRackauckas: add `saveat=t` here.

src/abc_inference.jl
```diff
@@ -0,0 +1,24 @@
...
+        randomized = VectorOfArray([sol(t[i]) for i in 1:length(t)])
```
ChrisRackauckas: with `saveat` set, you can avoid the interpolation and just use `convert(Array,sol)` directly.

ChrisRackauckas:
The code itself looks great! It's quite clean and I just have one small change request to avoid unnecessary interpolation.

However, your PR brought a bunch of temporary files along with it. It's not sufficient to just delete them either since then they will stay in your Git history. The easiest way to handle this is probably to just open a new PR and make sure you only include the correct files. I would highly recommend using a tool like GitKracken to help out.

PR #40, opened 2018-04-13
Added ABC inference using ApproxBayes.jl

marcjwilliams1:
Should be a clean PR without the tmp files.

### DIFF
--- a/src/abc_inference.jl
+++ b/src/abc_inference.jl
@@ -1,0 +1,23 @@
+function createabcfunction(prob, t, distancefunction, alg, kwargs...)
+    function simfunc(params, constants, targetdata)
+        sol = solve(problem_new_parameters(prob, params), alg, saveat = t, kwargs...)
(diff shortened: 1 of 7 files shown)


src/abc_inference.jl
```diff
@@ -0,0 +1,23 @@
+function createabcfunction(prob, t, distancefunction, alg, kwargs...)
+    function simfunc(params, constants, targetdata)
+        sol = solve(problem_new_parameters(prob, params), alg, saveat = t, kwargs...)
```
ChrisRackauckas: do `save_start = false`. Then you don't have to slice `[:, 2:end]` in the next line, and you'll save a matrix allocation.

ChrisRackauckas:
I made a minor change to cut out an allocation. Will merge when tests pass.

ChrisRackauckas:
Thank you for the contribution!

PR #42, opened 2018-05-03
RFC: Allow custom likelihood for dynamichmc_inference

mauro3:
This lets one choose a custom likelihood for `dynamichmc_inference`.  Not sure if this is in-line with the rest of the parameter-estimation functionality. 

TODO: add some docs.

### DIFF
--- a/src/abc_inference.jl
+++ b/src/abc_inference.jl
@@ -1,22 +1,22 @@
-function createabcfunction(prob, t, distancefunction, alg, kwargs...)
-    function simfunc(params, constants, targetdata)
-        sol = solve(problem_new_parameters(prob, params), alg, saveat = t, save_start = false, kwargs...)
-        data = convert(Array, sol)
-        distancefunction(targetdata, data), data
+function createabcfunction(prob, t, distancefunction, alg; kwargs...)
+    function simfunc(params, constants, data)
+        sol = solve(problem_new_parameters(prob, params), alg; saveat = t, save_start = false, kwargs...)
+        simdata = convert(Array, sol)
+        distancefunction(data, simdata), nothing
     end
 end
 
 function abc_inference(prob::DEProblem, alg, t, data, priors; ϵ=0.001,
-     distancefunction = euclidean, ABCalgorithm = ABCSMC, progress = false,
-     num_samples = 500, maxiterations = 10^5, kwargs...)
+                       distancefunction = euclidean, ABCalgorithm = ABCSMC, progress = false,
+                       num_samples = 500, maxiterations = 10^5, kwargs...)
 
-    abcsetup = ABCalgorithm(createabcfunction(prob, t, distancefunction, alg, kwargs...),
-     length(priors),
-     ϵ,
-     ApproxBayes.Prior(priors);
-     nparticles = num_samples,
-     maxiterations = maxiterations
-    )
+    abcsetup = ABCalgorithm(createabcfunction(prob, t, distancefunction, alg; kwargs...),
+                            length(priors),
+                            ϵ,
+                            ApproxBayes.Prior(priors);
+                            nparticles = num_samples,
+                            maxiterations = maxiterations
+                            )
 
     abcresult = runabc(abcsetup, data, progress = progress)
     return abcresult
--- a/src/dynamichmc_inference.jl
+++ b/src/dynamichmc_inference.jl
@@ -31,19 +31,13 @@ function (P::DynamicHMCPosterior)(θ)
 end
 
 function dynamichmc_inference(prob::DEProblem, alg, t, data, priors, transformations;
-                              σ=0.01, ϵ=0.001, initial=Float64[], iterations=1000, kwargs...)
+                              σ=0.01, ϵ=0.001, initial=Float64[], num_samples=1000, kwargs...)
     P = DynamicHMCPosterior(prob, data, priors, t, Normal(0.0, σ), alg, kwargs)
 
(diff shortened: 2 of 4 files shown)


ChrisRackauckas:
Yes, I think this fits in nicely with the other parameter estimation stuff, where for loss functions you can pass in `loss(sol)`. The Turing.jl handling should get similar functionality, Stan.jl handling requires Stan code (and has its own system here) so there's no use trying that. I would like to see something like this work with ABC too, and I presume it should work by just passing in a likelihood/loss function here: https://github.com/JuliaDiffEq/DiffEqBayes.jl/blob/master/src/abc_inference.jl#L4-L5

mauro3:
For Turing it would have to go here: https://github.com/JuliaDiffEq/DiffEqBayes.jl/blob/ecc1fa583ed3e7bbef9d98cafd980730c8cbed71/src/turing_inference.jl#L25, but there only a distribution works (I think).  So, that would be a bit inconsistent with DynamicHMC.

For ABC you propose to ditch the kwarg `distancefunction` and instead have it also as positional argument.  I can add this.

Not sure what the travis failure is about as it concerns ABC which I haven't touched yet.

mauro3:
I don't think the ABC interface fits into the scheme because the `simfunc` takes the data as an argument.  It would look something like:
```julia
function createabcfunction(prob, t, distancefunction, alg, kwargs...)
    function (params, constants, data)
        sol = solve(problem_new_parameters(prob, params), alg, saveat = t, save_start = false, kwargs...)
        simdata = convert(Array, sol)
        distancefunction(data, simdata), nothing
    end
end

function abc_inference(prob::DEProblem, alg, t, data, priors; ϵ=0.001,
                       distancefunction = euclidean, ABCalgorithm = ABCSMC, progress = false,
                       num_samples = 500, maxiterations = 10^5, kwargs...)

    abcfunction = createabcfunction(prob, t, distancefunction, alg; kwargs...)
    abc_inference(prob, alg, t, data, abcfunction, priors; ϵ=ϵ,
                  ABCalgorithm = ABCalgorithm, progress = progress,
                  num_samples = num_samples, maxiterations = maxiterations, kwargs...)
end

### This still carries `data`!:
function abc_inference(prob::DEProblem, alg, t, data, abcfunction, priors; ϵ=0.001,
                       ABCalgorithm = ABCSMC, progress = false,
                       num_samples = 500, maxiterations = 10^5, kwargs...)

    abcsetup = ABCalgorithm(abcfunction,
                            length(priors),
                            ϵ,
                            ApproxBayes.Prior(priors);
                            nparticles = num_samples,
                            maxiterations = maxiterations
                            )

    return runabc(abcsetup, data, progress = progress)
end
```

Instead I just added an example to the test of how to do a custom distance function (and fixed a bug with kwargs).

src/abc_inference.jl
```diff
@@ -1,22 +1,22 @@
-function createabcfunction(prob, t, distancefunction, alg, kwargs...)
-    function simfunc(params, constants, targetdata)
-        sol = solve(problem_new_parameters(prob, params), alg, saveat = t, save_start = false, kwargs...)
-        data = convert(Array, sol)
-        distancefunction(targetdata, data), data
+function createabcfunction(prob, t, distancefunction, alg; kwargs...)
+    function simfunc(params, constants, data)
+        sol = solve(problem_new_parameters(prob, params), alg; saveat = t, save_start = false, kwargs...)
```
mauro3: Using this with callbacks showed that setting the `saveat` option is very brittle as often other time-points are also saved (such as the time of the event).  I'll update it to just use interpolation.  Probably applies to other solvers too.

ChrisRackauckas: Yeah, that makes a lot of sense. These examples might work then if the user does `save_positions = (false,false)`. I think that an option to override callback saving may be worthwhile? I am not sure though. Another issue I've had with the current strategy is that it's not mutable. I think DiffEq.jl may need an issue on improving callback saving strategies.

mauro3: But for this I would, is should not matter too much to use interpolation, unless the solve is very cheap, right?

ChrisRackauckas: Yeah. The post-solution interpolation would just require some extra searches and have an extra allocation, but if the solve is expensive enough this wouldn't even be noticed and would be more robust. I can't say I've checked where the cutoff is, for example, what overhead would that add to Lotka-Volterra? No idea.

mauro3: Ok if I push a fix for this too to this PR?  I can do that now, and then we can merge.

mauro3: An I will squash into one commit.

Vaibhavdixit02: @mauro3 `abc_inference` doesn't face problems in case of callbacks, right?

ChrisRackauckas: I would like to get a sense of the cost from doing it first.

mauro3: Ok, let's merge it like this then and investigate the `saveat` in another PR.  In the problem I'm currently working with, which has callbacks, I had to modify it. This is what I use:
```julia
function createabcfunction(prob, t, distancefunction, alg; kwargs...)
    function simfunc(params, constants, data)
        sol = solve(problem_new_parameters(prob, params), alg; kwargs...)
        if any((s.retcode != :Success for s in sol))
            return Inf, nothing
        else
            simdata = convert(Array, sol(t))
            distancefunction(data, simdata), nothing
        end
    end
end
```

Maybe I ran the example in #43 with my fix, as Chris had ABC also fail there.

PR #54, opened 2018-07-17
Upgrade to 0.7

Vaibhavdixit02:


Vaibhavdixit02:
@ChrisRackauckas looks like Turing will need to be updated first?

ChrisRackauckas:
Yeah, it's likely that will be a holdup. Does it work other than that? You may need to disable the Turing part until they update. If it has the v0.7-beta2 requirement than it won't ruin v0.6 usage so it's okay to disable.

PR #57, opened 2018-09-16
Update REQUIRE

Vaibhavdixit02:


Vaibhavdixit02:
@ChrisRackauckas even though Turing 0.4.4 is being used (https://travis-ci.org/JuliaDiffEq/DiffEqBayes.jl/jobs/429234428#L1025) it trows error with Mamba and the stack trace (https://travis-ci.org/JuliaDiffEq/DiffEqBayes.jl/jobs/429234428#L1426) in the Travis mentions Line 24 here https://github.com/TuringLang/Turing.jl/blob/master/src/Turing.jl#L24 but it is empty now. It was the import statement for Mamba earlier. This looks like a case that some wrong version of Turing is being used by Travis?

ChrisRackauckas:
Why take off the lower bound? Lower bound it to 0.5.0.

Vaibhavdixit02:
0.5.0 is just a pre-release?

PR #58, opened 2018-09-17
compiling on v1.0

ChrisRackauckas:


ChrisRackauckas:
@Vaibhavdixit02

Vaibhavdixit02:
@xukai92 I am facing problem with adding Turing as a dependency in Julia 0.7, is this a known issue? I saw a 0.5.0 pre-release which is supposed to work on 1.0 but it is not available on 0.7?

REQUIRE
```diff
@@ -3,7 +3,7 @@ DiffEqBase 1.7.0
 #Mamba
 Stan
 Distributions
-Turing 0.5.0
+Turing 0.4.4
```
ChrisRackauckas: no, revert this.

ChrisRackauckas:
It was working fine locally. The issue may be the lower bound to the v0.4.0 + Pkg's enforcement of semvar. Don't make a lower bound lower than necessary: v0.5.0 of Turing.jl is required for v0.7.

Vaibhavdixit02:
v0.5.0 is not available on 0.7. Can you comfirm which version you used
locally?

ChrisRackauckas:
https://github.com/TuringLang/Turing.jl/blob/v0.5.0/REQUIRE

ChrisRackauckas:
https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/Turing/versions/0.5.0/requires

Issue #46, opened 2018-06-12
Update turing_inference to accommodate Turing's updated syntax.


Vaibhavdixit02:
Hi @xukai92 I saw that there have been quite a few changes in the last few months in Turing.jl and thus `turing_inference` needs to be updated too, could you help out with the required changes?

xukai92:
Sure, I will update Turing-related codes.

Vaibhavdixit02:
That would be a great help 👍

Vaibhavdixit02:
@xukai92 Could you provide a loose timeframe in which you will be able to take a look. I will be presenting a talk at JuliaCon'18 on DiffEqBayes.jl so it would be helpful for me to plan out my work for the talk accordingly. Thanks!

xukai92:
I will do it in the next two days.

xukai92:
I was trying to upgrade related packages for this repo. but failed to build `SymEngine`. 

```bash
=======================[ ERROR: SymEngine ]=======================

[...]

PR #68, opened 2019-02-09
Update turing_inference #46

Vaibhavdixit02:


Vaibhavdixit02:
@ChrisRackauckas this will not pass until a new release of Turing.jl, tested it locally with their master branch.

PR #75, opened 2019-06-21
generalize the likelihood definition in turing inference

ChrisRackauckas:


src/turing_inference.jl
```diff
@@ -1,16 +1,23 @@
 function turing_inference(prob::DiffEqBase.DEProblem,alg,t,data,priors = nothing;
+  likelihood_dist_priors = [InverseGamma(2, 3)],
+  likelihood = (u,p,t,σ) -> MvNormal(u, σ[1]*ones(length(u))),
```
Vaibhavdixit02: Only using ` σ[1]` here so why is  `σ` made a vector below?

ChrisRackauckas: Because that's just the default

PR #81, opened 2019-07-21
fix retcode handling

ChrisRackauckas:


src/turing_inference.jl
```diff
@@ -39,39 +39,32 @@ function turing_inference(prob::DiffEqBase.DEProblem,alg,t,data,priors;
...
-      res = sol_tmp.u
-      __lp = Turing.observe(
-        sampler,
-        likelihood(res,theta,Inf,σ),   # Distribution
-        x,    # Data point
-        vi
-      )
-      vi.logp += __lp
+    if any((s.retcode != :Success for s in sol_tmp)) && any((s.retcode != :Terminated for s in sol_tmp))
+      vi.logp += Inf
```
trappmartin: Should this not be `-Inf`? But I might easily misunderstand what you are trying to do.

PR #89, opened 2019-09-24
Update dynamic_inference

Vaibhavdixit02:
Allowing the user to pass a custom likelihood is not allowed right now, I'll add that with a follow up PR later.


src/DiffEqBayes.jl
```diff
@@ -20,7 +20,7 @@ function __init__()
     @require DynamicHMC="bbc10e6e-7c05-544b-b16e-64fede858acb" begin
         using DynamicHMC, TransformVariables, LogDensityProblems
         include("dynamichmc_inference.jl")
-        export dynamichmc_inference
+        export dynamichmc_inference, DynamicHMCPosterior
```
ChrisRackauckas: why export that?

PR #120, opened 2019-12-22
Isolate the Stan test group

ChrisRackauckas:


ChrisRackauckas:
@Vaibhavdixit02 anything you see here that would cause that execution time of the Stan tests to explode? Can you take a look?

PR #123, opened 2020-01-06
use Turing through imports

ChrisRackauckas:


Vaibhavdixit02:
Will require new release of AdvancedHMC https://github.com/TuringLang/AdvancedHMC.jl/pull/160, once it's in this should be good to go. Works for me locally.

xukai92:
The new release is there.

Vaibhavdixit02:
Thanks @xukai92!

ChrisRackauckas:
@xukai92 can we get some help figuring out what Turing is trying to iterate?

xukai92:
Do you mean why we do this DynamicPPL thing?

Vaibhavdixit02:
@ChrisRackauckas that happens because https://github.com/TuringLang/Turing.jl/blob/master/src/inference/Inference.jl#L856's types, the `MvNormal` is `<:` both `Distribution` and `MultivariateDistribution`, but the `MultivariateDistribution` method does not accept `AbstractArray` type, so when you don't matrixize in the `dot_tilde` call here, it can't iterate. On changing it though, I am encountering another error which I can't understand https://travis-ci.org/JuliaDiffEq/DiffEqBayes.jl/jobs/635608371?utm_medium=notification&utm_source=github_status

xukai92:
It's might be related to what @mohamed82008 is trying to fix.

mohamed82008:
Yes that looks related to https://github.com/TuringLang/Turing.jl/pull/1043. I will take a look next week. Please open an issue in Turing to track it, happy to investigate when I get some time.

Issue #125, opened 2020-01-22
Error when Turing 0.8.2 installed


jonasmac16:
Hi,
with `Turing` installed at version `0.8.2` in my Julia 1.3 env it get the following error running the example problem and it goes away once `Turing` is removed:
```
TypeError: in Type, in Targs, expected Targs<:NamedTuple, got Type{Tuple{:x}}
#turing_inference#1(::Array{InverseGamma{Float64},1}, ::DiffEqBayes.var"#8#17", ::Int64, ::NUTS{Turing.Core.ForwardDiffAD{40},(),AdvancedHMC.Adaptation.DiagEuclideanMetric}, ::Array{Symbol,1}, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(turing_inference), ::ODEProblem{Array{Float64,1},Tuple{Float64,Float64},true,Array{Float64,1},LotkaVolterraTest1{var"#40#41",Nothing,Nothing,Nothing,Nothing,Nothing,Any,Any},Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}},DiffEqBase.StandardODEProblem}, ::Tsit5, ::Array{Float64,1}, ::Array{Float64,2}, ::Array{Truncated{Normal{Float64},Continuous,Float64},1}) at turing_inference.jl:83
(::DiffEqBayes.var"#kw##turing_inference")(::NamedTuple{(:num_samples,),Tuple{Int64}}, ::typeof(turing_inference), ::ODEProblem{Array{Float64,1},Tuple{Float64,Float64},true,Array{Float64,1},LotkaVolterraTest1{var"#40#41",Nothing,Nothing,Nothing,Nothing,Nothing,Any,Any},Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}},DiffEqBase.StandardODEProblem}, ::Tsit5, ::Array{Float64,1}, ::Array{Float64,2}, ::Array{Truncated{Normal{Float64},Continuous,Float64},1}) at none:0
top-level scope at DiffEqBayes_issue.jl:26
 ```
```julia
using DiffEqBayes
using DifferentialEquations
using ParameterizedFunctions
using Distributions
using ForwardDiff

f1 = @ode_def_bare LotkaVolterraTest1 begin
 dx = a*x - x*y
 dy = -3*y + x*y
end a

[...]

PR #127, opened 2020-01-25
Fix Turing support

mohamed82008:
This PR makes DiffEqBayes and the latest Turing compatible and simplifies the Turing code in DiffEqBayes. Mamba seems to be causing a stackoverflow from the type piracy in https://github.com/brian-j-smith/Mamba.jl/blob/master/src/distributions/constructors.jl. I commented it out for now.

Relevant: #123, #125


ChrisRackauckas:
Awesome thanks!

PR #133, opened 2020-03-21
Remove Mamba dependency

### LINKED ISSUE #132: Remove Mamba dependency
Since #127 Mamba is no longer used, and the only place in the code-base [where I found it](https://github.com/JuliaDiffEq/DiffEqBayes.jl/search?q=Mamba&unscoped_q=Mamba) is in the stan_inference.jl [plotting functions](https://github.com/JuliaDiffEq/DiffEqBayes.jl/blob/afac88fd7f1bfbfee2cc8bf9dbfbb915b420b034/src/stan_inference.jl#L144-L153).

If the Mamba dependency is removed, I think https://github.com/JuliaDiffEq/DiffEqTutorials.jl/issues/129 could be more easily solved since Mamba is the only one (in DiffEqTutorials) that depends on Gadfly.

I can make a PR if you agree.

SebastianM-C:
Closes #132


SebastianM-C:
The test errors do not seem related to the PR (package resolution is fine).

PR #135, opened 2020-03-21
Use latest version of Turing

devmotion:
I just want to figure out what the issues exactly are.


ChrisRackauckas:
Oh, this is the ParameterizedFunction change https://github.com/JuliaDiffEq/ParameterizedFunctions.jl/pull/61 . I'll put a hotfix out on the Turing part. The Stan part I plan to get to later today. MATLABDiffEq is a similar issue (but this is why we're coalescing behind a single compiler backend now...)

PR #137, opened 2020-03-21
Remove plot_chain

devmotion:
This function does not seem to be very useful, but maybe there's a reason for having it in DiffEqBayes?


ChrisRackauckas:
Yeah, delete it. Over time the PPL converged on their plotting functions, so now it doesn't make any sense.

PR #138, opened 2020-03-22
update Stan inference code to use ModelingToolkit representation

ChrisRackauckas:


ChrisRackauckas:
Cool, the Stan code generator is fixed and now any Julia ODE problems can be used. Tests still fail though because of the chain stuff. @Vaibhavdixit02 can you do the finishing touches on this? I'll merge and tag because it's a big update and the last part is just fixing up the test, but it would be good to get that quick fix done ASAP

PR #139, opened 2020-03-22
update Stan getters

ChrisRackauckas:


ChrisRackauckas:
@Vaibhavdixit02 do you know why the Stan read_summary isn't found?

PR #140, opened 2020-03-23
reverse mode autodiff and more

AStupidBear:
- Support reverse mode autodiff for Turing and DynamicHMC
- Add a `obsvbls` kwarg to `stan_inference`, `dynamichmc_inference` and `turing_inference` for partially observable data
- Add a `sample_u0` kwarg to `stan_inference`, `dynamichmc_inference` and `turing_inference` for joint inference of `u0` and `p`
- Add a `diffeq_string` kwarg to `stan_inference` for complex ODE specification


src/dynamichmc_inference.jl
```diff
@@ -25,16 +27,28 @@ Base.@kwdef struct DynamicHMCPosterior{TA,TP,TD,TT,TR,TS,TK}
...
+    p = convert.(T, sample_u0 ? parameters[(nu + 1):end] : parameters)
+    _saveat = t === nothing ? Float64[] : t
+    prob = remake(problem, u0 = Tracker.data.(u0), p = Tracker.data.(p))
+    solution = solve(prob, algorithm; saveat = _saveat, solve_kwargs...)
+    if T <: Tracker.TrackedReal || adkind == Val(:Zygote)
+        solution′ = concrete_solve(problem, algorithm, u0, p; saveat = _saveat, solve_kwargs...)
+    else
+        solution′ = solution
+    end
+    any((s.retcode ≠ :Success && s.retcode ≠ :Terminated) for s in solution) && return T(0) * sum(σ) + T(-1e10)
```
ChrisRackauckas: why `-1e10`?

src/dynamichmc_inference.jl
```diff
@@ -25,16 +27,28 @@ Base.@kwdef struct DynamicHMCPosterior{TA,TP,TD,TT,TR,TS,TK}
...
-    log_likelihood = sum(sum(logpdf.(Normal.(0.0, σ), solution(t) .- data[:, i]))
-                         for (i, t) in enumerate(t))
+    @unpack algorithm, problem, data, t, parameter_priors = P
+    @unpack σ_priors, solve_kwargs, obsvbls, sample_u0, adkind = P
+    nu, T = length(problem.u0), eltype(parameters)
+    u0 = convert.(T, sample_u0 ? parameters[1:nu] : problem.u0)
+    p = convert.(T, sample_u0 ? parameters[(nu + 1):end] : parameters)
+    _saveat = t === nothing ? Float64[] : t
+    prob = remake(problem, u0 = Tracker.data.(u0), p = Tracker.data.(p))
+    solution = solve(prob, algorithm; saveat = _saveat, solve_kwargs...)
```
ChrisRackauckas: this will solve twice if Zygote: you might want to just have it be concrete_solve.

src/stan_inference.jl
```diff
@@ -49,11 +48,16 @@ end
 function stan_inference(prob::DiffEqBase.DEProblem,t,data,priors = nothing;alg=:rk45,
                             num_samples=1000, num_warmup=1000, reltol=1e-3,
                             abstol=1e-6, maxiter=Int(1e5),likelihood=Normal,
-                            vars=(StanODEData(),InverseGamma(3,3)),nchains=1)
+                            vars=(StanODEData(),InverseGamma(3,3)),nchains=1,
+                            obsvbls = 1:size(data, 1), sample_u0 = false, diffeq_string = "")
```
ChrisRackauckas: what's the application for having `diffeq_string` here? Just curious. Might be easier to have a default of `nothing`.

src/turing_inference.jl
```diff
@@ -8,7 +10,10 @@ function turing_inference(
     likelihood = (u,p,t,σ) -> MvNormal(u, σ[1]*ones(length(u))),
     num_samples=1000, sampler = Turing.NUTS(0.65),
     syms = [Turing.@varname(theta[i]) for i in 1:length(priors)],
-    progress = false, kwargs...,
+    obsvbls = 1:size(data, 1),
```
ChrisRackauckas: it would be better to just use `saveidxs` here, since that would reduce the memory use.

src/dynamichmc_inference.jl
AStupidBear: Because some optimization packages such as `Optim.jl` won't accept `Inf` values ( `Optim` is used in `DynamicHMC.jl`'s warmup phases). `T(0) * sum(σ)` is used because `Tracker` will error if the output is constant and `Zygote` will give a `nothing` for its gradient which is incompatible with the entire pipeline.

Optimization can be used for MAP too. 
```
function get_nlogp(model)
    vi = Turing.VarInfo(model)
    function nlogp(θ)
        spl = Turing.SampleFromPrior()
        vi′ = Turing.VarInfo(vi, spl, θ)
        model(vi′, spl)
        -Turing.getlogp(vi′)
    end
    back(θ) = l̄ -> (-Turing.gradient_logp_reverse(θ, vi, model)[2],)
    @eval Zygote.@adjoint $nlogp(θ) = $nlogp(θ), $back(θ)
    return nlogp
end
res_turing = turing_inference(...)
)
nlogp = get_nlogp(res_turing.info.model)
DiffEqFlux.sciml_train(nlogp, ...)
```

src/dynamichmc_inference.jl
AStupidBear: Yes. If we just keep `concrete_solve`, we won't be able to access the return code of a solution. So should we modify `concrete_solve` to return `solution` + `retcode`? Or we just infer the failure from the shape of `sol`? If `size(sol, 2) < length(saveat)`, then there's a failure?

src/stan_inference.jl
AStupidBear: I will make `diffeq_string` default to `nothing`. This is my use case.

```
diffeq_string = """
real af(real x, real theta, real slope) {
    return 1 / (1 + exp(-(x - theta) / slope));
}
real[] sho(real t, real[] u, real[] p, real[] x_r, int[] x_i) {
    real du[12];
    real x1 = u[1]; real y1 = u[2]; real x2 = u[3]; real y2 = u[4]; real m1 = u[5]; real k1 = u[6];
    real x3 = u[7]; real y3 = u[8]; real x4 = u[9]; real y4 = u[10]; real  m2 = u[11]; real k2 = u[12];
    real r = p[1]; real s1 = p[2]; real s2 = p[3]; real eps = p[4]; real eta = p[5]; real w_12 = p[6];
    real w_21 = p[7]; real theta = p[8]; real v_syn = p[9]; real wk1 = p[10]; real g1 = p[11]; real g2 = p[12];
    real tau = 0.05; real taum = 0.2; real tau2 = 0.5; real tauk = 0.5;
    real wk2 = 1; real w_34 = 0; real w_43 = 0; real eta2 = 6;
    real W = p[12 + bin_search(ceil((t - 1) / $dt), 0, 1000)];
    du[1] = (r + W - s2 * x1^3 + s1 * x1 - eps * y1 - w_12 * af(x2, theta, 0.1) * (x1 - v_syn) + g1 * (x3 - x1) - eta * k1) / tau;
    du[2] = -y1 + af(x1, 0, 0.5);
    du[3] = (r + W - s2 * x2^3 + s1 * x2 - eps * y2 - w_21 * af(x1, theta, 0.1) * (x2 - v_syn) + g2 * (x4 - x2) + eta * k1) / tau;
    du[4] = -y2 + af(x2, 0, 0.5);
    du[5] = (-m1 + wk1 * (af(x1, theta, 0.1) - af(x2, theta, 0.1))) / taum;
    du[6] = (-k1 + m1) / tauk;
    du[7] = (x3 - x3^3 - w_34 * af(x4, theta, 0.1) * (x3 - v_syn) + g1 * (x1 - x3) - eta2 * k2) / tau2;
    du[8] = -y3 + af(x3, 0, 0.5);
    du[9] = (x4 - x4^3 - w_43 * af(x3, theta, 0.1) * (x4 - v_syn) + g2 * (x2 - x4) + eta2 * k2) / tau2;
    du[10] = -y4 + af(x4, 0, 0.5);
    du[11] = (-m2 + wk2 * (af(x3, theta, 0.1) - af(x4, theta, 0.1))) / taum;
    du[12] = (-k2 + m2) / tauk;
    return du;
}
"""
res_stan = stan_inference(
    prob, (0:dt:tend) .+ 1, data, priors,
    num_samples = 5000, num_warmup = 500,
    vars =(DiffEqBayes.StanODEData(), InverseGamma(3, 2)), 
    obsvbls = 6:6, sample_u0 = true, diffeq_string = diffeq_string
)
```

`real W = p[12 + bin_search(ceil((t - 1) / $dt), 0, 1000)];` cannot be handled easily in Julia.

src/turing_inference.jl
AStupidBear: Will do that.

src/dynamichmc_inference.jl
AStupidBear: It seems that `size(sol, 2) < length(_saveat)` requires much less better.

src/stan_inference.jl
ChrisRackauckas: yeah, interesting. The new ModelingToolkit form will handle the embded function, but it won't call the Stan `bin_search` function.

src/dynamichmc_inference.jl
ChrisRackauckas: Optim accepts and works with Infs just fine by rejecting those steps. Using a largeish value might be the cause of the test failures?

ChrisRackauckas:
Looks good to me, but tests are failing so that's a bad sign.

ChrisRackauckas:
Thanks a ton, looks great.

PR #146, opened 2020-04-05
Add tests for save_idxs and update readme to current api

abhigupta768:


Vaibhavdixit02:
Add a test for `sample_u0` as well for all the backends, that isn't tested yet

Vaibhavdixit02:
Looks good to me now.

@ChrisRackauckas?

src/dynamichmc_inference.jl
```diff
@@ -28,17 +28,19 @@ Base.@kwdef struct DynamicHMCPosterior{TA,TP,TD,TT,TR,TS,TK}
...
 end
 
 function (P::DynamicHMCPosterior)(θ)
     @unpack parameters, σ = θ
     @unpack algorithm, problem, data, t, parameter_priors = P
-    @unpack σ_priors, solve_kwargs, sample_u0 = P
-    nu, T = length(problem.u0), eltype(parameters)
+    @unpack σ_priors, solve_kwargs, sample_u0, save_idxs = P
+    T = eltype(parameters)
+    nu = save_idxs === nothing ? length(problem.u0) : length(save_idxs)
```
Vaibhavdixit02, line 39: Revert this back and let's handle the combination from a different PR.

src/turing_inference.jl
```diff
@@ -30,7 +30,7 @@ function turing_inference(
         for i in 1:length(likelihood_dist_priors)
             σ[i] ~ likelihood_dist_priors[i]
         end
-        nu = length(prob.u0)
+        nu = save_idxs === nothing ? length(prob.u0) : length(save_idxs)
```
Vaibhavdixit02, line 33: Same

README.md
```diff
@@ -126,7 +137,7 @@ equation solver.
 
  bayesian_result_abc = abc_inference(prob1, Tsit5(), t, data, priors)
 ```
-### Using save_idxs for using latent variables 
+### Using save_idxs to declare observables
```
ChrisRackauckas, line 141: you might want text here. Describe what the problem is, how to fix it, then show code.

Vaibhavdixit02:
@abhigupta768 for `sample_u0` to be actually usable in real problems I think it is very important to have it be compatible with `save_idxs`. Let's follow up with a PR to fix that.

PR #159, opened 2020-05-09
Update to Turing 0.12

devmotion:
This is an alternative to https://github.com/SciML/DiffEqBayes.jl/pull/158 that passes tests locally.

### DIFF
--- a/src/turing_inference.jl
+++ b/src/turing_inference.jl
@@ -44,7 +44,8 @@ function turing_inference(
         failure = size(sol, 2) < length(_saveat)
 
         if failure
-            @logpdf() = T(0) * sum(x) + T(-Inf)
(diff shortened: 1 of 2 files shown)


src/turing_inference.jl
```diff
@@ -44,7 +44,8 @@ function turing_inference(
         failure = size(sol, 2) < length(_saveat)
 
         if failure
-            @logpdf() = T(0) * sum(x) + T(-Inf)
+            S = typeof(Turing.Inference.getlogp(_varinfo))
+            Turing.Inference.acclogp!(_varinfo, S(-Inf))
```
ChrisRackauckas, line 48: I think the other parts were just there so the output type would be stable.

PR #161, opened 2020-05-09
Remove warning about internal variables

devmotion:
Turing now supports the syntax discussed in https://github.com/SciML/DiffEqBayes.jl/pull/158#issuecomment-626139149 to eliminate the warning when using internal variables.

### DIFF
--- a/src/turing_inference.jl
+++ b/src/turing_inference.jl
@@ -21,7 +21,7 @@ function turing_inference(
     kwargs...,
 )
     N = length(priors)
-    Turing.@model mf(x, ::Type{T} = Float64) where {T <: Real} = begin
+    Turing.@model function mf(x, ::Type{T} = Float64) where {T <: Real}
         theta = Vector{T}(undef, length(priors))
         for i in 1:length(priors)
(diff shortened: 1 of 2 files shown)


src/turing_inference.jl
```diff
@@ -39,13 +39,12 @@ function turing_inference(
...
         end
-        _saveat = isnothing(t) ? Float64[] : t
+        _saveat = t === nothing ? Float64[] : t
         sol = concrete_solve(prob, alg, u0, p; saveat = _saveat, progress = progress, save_idxs = save_idxs, kwargs...)
         failure = size(sol, 2) < length(_saveat)
 
         if failure
-            S = typeof(Turing.Inference.getlogp(_varinfo))
-            Turing.Inference.acclogp!(_varinfo, S(-Inf))
+            Turing.acclogp!(_varinfo, -Inf)
```
devmotion, line 47: Turing (or rather DynamicPPL) will call `S(-Inf)` internally to convert it to the type of `getlogp(_varinfo)`, so IMO DiffEqBayes should not have to care about it.

ChrisRackauckas: Alright

PR #168, opened 2020-05-24
Small update of stan_inference.jl to include Stan's exposed Adams solver

wallyxie:
Stan has had an Adams-Moulton method exposed for the past couple of years now as `integrate_ode_adams`, but I only recently found out because it was only explicitly mentioned in the latest update of their documentation. Refer to https://github.com/stan-dev/stan/issues/2542. Now recommended as the default non-stiff solver by people in the Stan community over rk45. Just a small suggested update to accommodate that as a result. Thanks to the DifferentialEquations.jl team for their hard work once more!

### DIFF
--- a/src/stan_inference.jl
+++ b/src/stan_inference.jl
@@ -47,7 +47,7 @@ function generate_theta(n,priors)
 end
 
 function stan_inference(prob::DiffEqBase.DEProblem,t,data,priors = nothing,
-                            stanmodel = nothing;alg=:rk45,
+                            stanmodel = nothing;alg=:adams,
                             num_samples=1000, num_warmup=1000, reltol=1e-3,
                             abstol=1e-6, maxiter=Int(1e5),likelihood=Normal,
                             vars=(StanODEData(),InverseGamma(3,3)),nchains=1,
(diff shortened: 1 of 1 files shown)


src/stan_inference.jl
```diff
@@ -47,7 +47,7 @@ function generate_theta(n,priors)
 end
 
 function stan_inference(prob::DiffEqBase.DEProblem,t,data,priors = nothing,
-                            stanmodel = nothing;alg=:rk45,
+                            stanmodel = nothing;alg=:adams,
```
ChrisRackauckas, line 50:
```suggestion
                            stanmodel = nothing;alg=:rk45,
```

I wouldn't make it the default because if the rk45 is any good it should perform better on a lot of standard non-stiff problems.

ChrisRackauckas:
Thanks!

src/stan_inference.jl
wallyxie: Yep, sounds good!

Vaibhavdixit02:
The error is unrelated and will be fixed with the new release of DiffEqSensitivity

PR #184, opened 2020-11-15
Update make.jl

mkg33:
Added URL and authors.


docs/make.jl
```diff
@@ -2,13 +2,13 @@ using Documenter, DiffEqBayes
 
 makedocs(
     sitename="DiffEqBayes.jl",
-    authors="#",
+    authors="Chris Rackauckas et al.",
```
ChrisRackauckas, line 5: Add @Vaibhavdixit02 in here.

PR #244, opened 2022-06-18
Bump versions and some deprecations updates

Vaibhavdixit02:


Project.toml
```diff
@@ -34,25 +34,25 @@ Turing = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
...
 RecursiveArrayTools = "1,2"
 Reexport = "0.2, 1.0"
 Requires = "0.5, 1.0"
-StanSample = "3, 4"
+StanSample = "3, 4, 6"
 StructArrays = "0.4, 0.5, 0.6"
-TransformVariables = "0.3, 0.4"
-Turing = "0.12, 0.13, 0.14, 0.15, 0.16, 0.17"
+TransformVariables = "0.3, 0.4, 0.6"
+Turing = "0.12, 0.13, 0.14, 0.15, 0.16, 0.17, 0.21"
```
ChrisRackauckas, line 55: We should drop some old versions here since many of those wouldn't ever be used on v1.6

PR #258, opened 2022-09-06
add delta key word

EvoArt:
To allow user to  modify, adapt_delta parameter, e.g. in Turing this is often set lower for ODE models.

### DIFF
--- a/src/stan_inference.jl
+++ b/src/stan_inference.jl
@@ -62,6 +62,7 @@ function stan_inference(prob::DiffEqBase.DEProblem,
                         # stan_sample keyword arguments
                         num_samples = 1000, num_warmups = 1000,
                         num_cpp_chains = 1, num_chains = 1, num_threads = 1,
+                        delta=0.8,
                         # read_samples arguments
                         output_format = :mcmcchains,
                         # read_summary arguments
@@ -200,7 +201,7 @@ function stan_inference(prob::DiffEqBase.DEProblem,
                 "t0" => prob.tspan[1], "ts" => t)
 
     @time rc = stan_sample(stanmodel; data, num_threads, num_cpp_chains,
-                           num_samples, num_warmups, num_chains)
+                           num_samples, num_warmups, num_chains,delta)
 
     if success(rc)
         return StanResult(stanmodel, rc, read_samples(stanmodel, output_format))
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Run the formatter

src/stan_inference.jl
```diff
@@ -62,6 +62,7 @@ function stan_inference(prob::DiffEqBase.DEProblem,
                         # stan_sample keyword arguments
                         num_samples = 1000, num_warmups = 1000,
                         num_cpp_chains = 1, num_chains = 1, num_threads = 1,
+                        delta=0.8,
```
devmotion, line 65: Is that the default value if it is not specified?

src/stan_inference.jl
```diff
@@ -200,7 +201,7 @@ function stan_inference(prob::DiffEqBase.DEProblem,
                 "t0" => prob.tspan[1], "ts" => t)
 
     @time rc = stan_sample(stanmodel; data, num_threads, num_cpp_chains,
-                           num_samples, num_warmups, num_chains)
+                           num_samples, num_warmups, num_chains,delta)
```
devmotion, line 204:
```suggestion
                           num_samples, num_warmups, num_chains, delta)
```

src/stan_inference.jl
devmotion: To make the formatter happy (I guess):
```suggestion
                        delta = 0.8,
```

ChrisRackauckas:
Test failure looked unrelated. I turned off fail-fast so that the Stan tests will finish.

src/stan_inference.jl
```diff
@@ -62,6 +62,7 @@ function stan_inference(prob::DiffEqBase.DEProblem,
                         # stan_sample keyword arguments
                         num_samples = 1000, num_warmups = 1000,
                         num_cpp_chains = 1, num_chains = 1, num_threads = 1,
+                        delta = 0.8,
```
devmotion, line 65: On a second thought, I wonder if it would be easier and possibly better to group all `stan_sample` keyword arguments in a `stan_kwargs`, similar to `mcmc_kwargs` (https://github.com/SciML/DiffEqBayes.jl/blob/a1c6b90ba0f4dfedc4e41fc56212709b254ecf19/src/dynamichmc_inference.jl#L103), and then just forward it to `stan_sample`? Or even just forward all `kwargs...`?

Then it would be easier to customize also other Stan settings.

ChrisRackauckas: yeah that would probably be best

ChrisRackauckas:
We can generalize this in a follow-up. I agree that a named tuple approach is probably better

PR #259, opened 2022-09-06
Account for changes in DynamicHMC 3.3.0

devmotion:
It seems in DynamicHMC >= 3.3.0 `mcmc_with_warmup` returns a NamedTuple with field `posterior_matrix` whereas in older versions it is called `chain`.


ChrisRackauckas (approved):

ChrisRackauckas:
The test failure looks real

devmotion:
Yep. At first glance it seemed only the name was changed but I guess I have to check more carefully if there were other internal changes.

Issue #261, opened 2022-09-29
Duplicate using statement and formatting


AP6YC:
I realize that this is perhaps the most insignificant issue ever, but for the sake of code style I noticed that there is a duplicate `using DocStringExtensions` statement in `DiffEqBayes.jl` (I realize that duplicate imports don't change functionality).

While talking about `using` statements, I think that it is also worthwhile in general to break apart these statements into new lines to reduce git noise. If agreed, what is the preferred style of the authors and Julia packages in general?

With a single comma-separated statement with line breaks?
```julia
using
    DiffEqBase,
    Distributions,
    Turing,
    ...
```

Or multiple using statements?
```julia
using DiffEqBase
using Distributions
using Turing
...
```

Or something else?

PR #262, opened 2022-09-29
Feature/using-refactor

AP6YC:
This PR addresses my extremely pedantic minor issue #261, breaking apart the using statements and removing duplicates.

I followed the [BlueStyle](https://github.com/invenia/BlueStyle#module-imports) since the [YAS Style](https://github.com/jrevels/YASGuide) and [Julia Formatter](https://domluna.github.io/JuliaFormatter.jl/dev/style/) don't make much mention of this sort of thing.

### DIFF
--- a/src/DiffEqBayes.jl
+++ b/src/DiffEqBayes.jl
@@ -7,8 +7,6 @@ module DiffEqBayes
 using DiffEqBase
 using Distances
 using Distributions
-using Distributions
-using DocStringExtensions
 using DocStringExtensions
 using LinearAlgebra
 using MacroTools
--- a/src/DiffEqBayes.jl
+++ b/src/DiffEqBayes.jl
@@ -4,10 +4,21 @@ $(DocStringExtensions.README)
 module DiffEqBayes
 
 using DocStringExtensions
-using DiffEqBase, Distributions, Turing, MacroTools
-using RecursiveArrayTools, ModelingToolkit, LinearAlgebra
-using Parameters, Distributions, Optim, Requires
-using Distances, DocStringExtensions, Random, StanSample
+using DiffEqBase
+using Distributions
+using Turing
+using MacroTools
+using RecursiveArrayTools
+using ModelingToolkit
+using LinearAlgebra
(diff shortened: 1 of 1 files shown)


Vaibhavdixit02:
Indeed it looks like we don't cover this point in SciMLStyle https://github.com/SciML/SciMLStyle#modules. I personally prefer having the using statements bunch up multiple packages such that those packages have some similar use in the loading package, but that said I don't have any issue with these separate using statements

ChrisRackauckas:
Please revert. Issues of style should be settled in the style repo. And in any case, there is a clear answer here. It should use commas instead of separate usings because there are multi load optimizations in the works which would allow for invalidations to be only checked at the end of a full using statement. Therefore in the near future we may see the comma version as having a much lower load time. 

Then, in our style guide we always prefer using space instead of new lines for everything. For example, when argument numerous enough to fill more than a line, we keep things to the standard line length and then fill each subsequent line. We very specifically note that it is bad style to put one argument per line because it can make code hard to read. So extrapolating to this case, we shouldn't be doing one package per line.

We should clarify this in the style guide. And we should remove the duplicate using here. But generally the style change should be reverted.

Sorry about that, I just didn't get a chance to respond

AP6YC:
@ChrisRackauckas Thanks so much for the reply. It was absolutely my error to not see the SciML style repo that articulates your points. Your points are very illuminating (especially about multi load optimizations, which I look forward to reading about).

@Vaibhavdixit02 I see the reasoning in grouping packages with similar use, especially in consideration of the style Chris lays out. Sorry to make more work for you in reverting the PR. I'll do more due diligence next time.

Vaibhavdixit02:
Not at all @AP6YC, it brought up an interesting point. Please feel free to comment on the issue I created in SciMLStyle if you have anything more in mind for this

ChrisRackauckas:
Added to the style guide https://github.com/SciML/SciMLStyle/commit/60e2ba9fdc83aea3a82887fdc319c624e525ac71 thanks!

PR #274, opened 2022-12-10
Build doc examples as strict

ChrisRackauckas:


ChrisRackauckas:
@Vaibhavdixit02 can you take this from here? DiffEqBayes and DiffEqParamEstim I think are the last libraries with examples that aren't being build tested, so I'd hope to clean that up by the end of the week.

Vaibhavdixit02:
Sure

PR #289, opened 2022-12-12
use normal priors in pendulum example, to avoid stepsize bisection intermittent failure

Vaibhavdixit02:


docs/src/examples/pendulum.md
```diff
@@ -73,7 +73,7 @@ information, but have a prior belief that ω is between 0.1 and 3.0, while the
 length of the pendulum L is probably around 3.0:
 
 ```@example pendulum
-priors = [Uniform(0.1,3.0), Normal(3.0,1.0)]
+priors = [Normal(0.1,1.0), Normal(3.0,1.0)]
```
ChrisRackauckas, line 76: that's going to have a hard time staying positive.

PR #304, opened 2023-05-12
Prepare handling of truncated distribution for changes in Distributions

devmotion:
These changes are in preparation for https://github.com/JuliaStats/Distributions.jl/pull/1720.

### DIFF
--- a/src/stan_string.jl
+++ b/src/stan_string.jl
@@ -207,9 +207,11 @@ function stan_string(p::Union{Type{VonMises}, VonMises})
         return string("von_mises")
     end
 end
-function stan_string(p::T) where {T <: Truncated}
-    lower = isinf(p.lower) ? "" : p.lower
-    upper = isinf(p.upper) ? "" : p.upper
+function stan_string(p::Truncated)
+    p_lower = p.lower
+    p_upper = p.upper
+    lower = p_lower === nothing || isinf(p_lower) ? "" : p_lower
+    upper = p_upper === nothing || isinf(p_upper) ? "" : p_upper
     raw_string = stan_string(p.untruncated)
     return string(raw_string, " T[$lower,$upper]")
 end
--- a/src/stan_string.jl
+++ b/src/stan_string.jl
@@ -208,11 +208,10 @@ function stan_string(p::Union{Type{VonMises}, VonMises})
     end
 end
 function stan_string(p::Truncated)
-    p_lower = p.lower
-    p_upper = p.upper
+    min_truncated, max_truncated = extrema(p)
     min_untruncated, max_untruncated = extrema(p.untruncated)
-    lower = p_lower === nothing || p_lower == min_untruncated ? "" : string(p_lower)
-    upper = p_upper === nothing || p_upper == max_untruncated ? "" : string(p_upper)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
This will need a test and a lower bound?

devmotion:
Not sure about a test but this PR is backwards compatible, so no lower bound is needed. The checks `=== nothing` will just always evaluate to `false` in current and previous versions of Distributions. Only once the PR in Distributions is released, it can evaluate to `true` in some cases.

src/stan_string.jl
```diff
@@ -207,9 +207,11 @@ function stan_string(p::Union{Type{VonMises}, VonMises})
...
     end
 end
-function stan_string(p::T) where {T <: Truncated}
-    lower = isinf(p.lower) ? "" : p.lower
-    upper = isinf(p.upper) ? "" : p.upper
+function stan_string(p::Truncated)
+    p_lower = p.lower
+    p_upper = p.upper
+    lower = p_lower === nothing || isinf(p_lower) ? "" : p_lower
+    upper = p_upper === nothing || isinf(p_upper) ? "" : p_upper
```
ararslan, line 214:
```suggestion
    p_lower, p_upper = extrema(p)
    lower = isinf(p_lower) ? "" : p_lower
    upper = isinf(p_upper) ? "" : p_upper
```
Couldn't you use `extrema` and also be completely backwards-compatible?

devmotion: I could but here in DiffEqBayes it would be advantageous to know more: We're after whether the distribution is truncated with a lower and/or upper bound, and the `nothing` check is the easiest way to obtain this information in future versions of Distributions. `extrema` would require some additional comparisons with the `extrema` of the untruncated distribution - probably this should be done instead of checking `isinf` on both older and newer versions of Distributions, I just did not want to change the behaviour with released versions of Distributions in this PR. But maybe I should have done that:

```suggestion
    p_lower = p.lower
    p_upper = p.upper
    min_untruncated, max_untruncated = extrema(p.untruncated)
    lower = p_lower === nothing || p_lower == min_untruncated ? "" : string(p_lower)
    upper = p_upper === nothing || p_upper == max_untruncated ? "" : string(p_upper)
```

Of course, one could avoid the `nothing` check by writing
```suggestion
    p_lower, p_upper = extrema(p)
    min_untruncated, max_untruncated = extrema(p.untruncated)
    lower = p_lower === nothing || p_lower == min_untruncated ? "" : string(p_lower)
    upper = p_upper === nothing || p_upper == max_untruncated ? "" : string(p_upper)
```
instead. But this leads to less efficient code if a bound is `nothing`, and it is also not much safer API-wise since it still requires to access `p.untruncated`.

devmotion: I changed it to the version with `extrema` since it seems safer for the check: when calling `extrema` the returned bounds are ensured to be between the bounds of the untruncated distribution, but this is not guaranteed (even currently) if the fields are accessed directly.

devmotion:
I added some tests in test/stan_string.jl since it seems tests for `Truncated` distributions were missing.

Vaibhavdixit02 (approved):
Looks good, thanks!

PR #381, opened 2026-06-06
Re-enable downgrade CI strict (natural RED pending Turing/DynamicPPL #370)

ChrisRackauckas-Claude:
Re-enables the downgrade workflow (was `if: false`, ref #370) and makes it run **strict** so it fails **naturally** on the real upstream blocker, rather than being skipped or engineered green.

## Approach: run strict, expect a natural RED (not a skipped/disabled job, not a forced pass)

- `.github/workflows/Downgrade.yml`: removed `if: false` so the job **runs**. Kept `group: "Core"` (the `Stan` group needs CmdStan system deps the shared downgrade workflow does not install, so it would fail for the wrong reason). Kept `julia-version: "1.10"`.
- **Strict resolution** is in force: the `SciML/.github` `@v1` reusable downgrade workflow now **hardcodes `allow_reresolve=false`** and removed the `allow-reresolve` input (upstream change 2026-06-07). Passing that input from the caller is now an unknown input and causes a `startup_failure` instead of a real run, so the input is intentionally **not** passed. Strict is preserved by the reusable workflow default.
- `Project.toml [compat]`: reverted the previously raised floors back to **master's original lower bounds**. Floors are not lowered below master and not raised to force a pass.

## Why RED is expected
At master's floors under strict resolution on Julia 1.10, the Turing/DynamicPPL stack does not resolve/test cleanly — the known upstream blocker tracked in #370. The red X is the correct, visible signal; it is intentional and preferable to a disabled (`if: false`) job that hides the regression. **It will auto-green once Turing/DynamicPPL (#370) is resolved.**

> Ignore until reviewed by @ChrisRackauckas.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ChrisRackauckas-Claude:
Pushed a follow-up commit removing the now-invalid `allow-reresolve:` input from `Downgrade.yml`. The reusable `SciML/.github` `downgrade.yml` workflow is being changed to always use `allow_reresolve: false` and to drop the `allow-reresolve` `workflow_call` input entirely (see SciML/.github PR). Once that .github PR merges and `@v1` is retagged, any caller still passing `allow-reresolve:` would error with an invalid-input failure, so this line must go. Behavior is unchanged (it was already strict/false here).

ChrisRackauckas-Claude:
Pushed a corrective commit (6ab3a4f) to switch this PR from "engineer the downgrade green" to "run it strict and let it fail naturally":

- Re-added `allow-reresolve: false` (strict) to `Downgrade.yml` and dropped the `group: "Core"` filter so it matches master's strict config. The job is no longer `if: false`, so it runs and is visibly red.
- Reverted the `[compat]` floors back to master's original lower bounds (no longer raised to force a pass; not lowered below master either). `julia` stays at `1.10`.

Net effect: the downgrade job RUNS strict and is intentionally RED — a natural failure, not a disabled/skipped job — pending the upstream Turing/DynamicPPL blocker (#370). It will auto-green when that resolves.

> Ignore until reviewed by @ChrisRackauckas.

ChrisRackauckas-Claude:
Update: the strict config is now achieved correctly via the upstream default rather than by passing an input.

The `SciML/.github` `@v1` reusable downgrade workflow removed the `allow-reresolve` input on 2026-06-07 and now hardcodes `allow_reresolve=false`. Re-adding `allow-reresolve: false` on the caller side produced a `startup_failure` (unknown input) twice — i.e. the workflow couldn't even start, which is NOT the natural #370 failure we want. Final state (commit 772b21c):

- `Downgrade.yml`: no `if: false`; no `allow-reresolve` input (strict is the reusable workflow's hardcoded default); `group: "Core"`; `julia-version: "1.10"`.
- `Project.toml [compat]`: floors reverted to master's original lower bounds (not raised, not lowered).

Net: the downgrade job RUNS strict and is intentionally RED on the Turing/DynamicPPL blocker (#370) — a natural failure, not a disabled/skipped job or a forced green. Auto-greens when #370 resolves.

> Ignore until reviewed by @ChrisRackauckas.

PR #389, opened 2026-06-19
Raise stale downgrade-compat floors to fix Downgrade CI

ChrisRackauckas-Claude:
## Problem

The default-branch (`master`) CI is red on **Downgrade / Downgrade Tests - Core**. That job resolves the merged test environment with every `[compat]` entry pinned to its floor (`julia-downgrade-compat`, `--min=@deps`, Julia 1.10). The resolution fails with `Unsatisfiable`.

## Root cause

Many `[compat]` lower bounds are stale relative to the modern dependency generation the package actually targets (`RecursiveArrayTools = "4.3"`, `ModelingToolkit = "11.26"`, `Turing = "0.45"`, plus `DiffEqBase`/`OrdinaryDiffEq` 7.x). When pinned to their floors, those old lower bounds are mutually unsatisfiable with the high floors — e.g. `OrdinaryDiffEq = "6, 7"` pinned to 6.0.0 can never coexist with `RecursiveArrayTools = "4.3"` (RAT 4.x requires ODE 7), `TransformedLogDensities = "1"` (floor 1.0.0) requires `TransformVariables = "0.6"` while the repo pins `TransformVariables = "0.8"`, etc. This is a chain of ~21 stale floors.

## Fix

Raised each conflicting floor to the lowest registry version compatible with the rest of the pinned floors. Raising a lower bound is always safe for the normal resolution (it still selects the latest versions). Core SciML floors (`ModelingToolkit`, `RecursiveArrayTools`, `Turing`, `OrdinaryDiffEq`-as-`7`) were left as-is wherever possible; they only ever appeared as *demanders* of higher versions, never as the binding lower cap.

| Package | Old | New |
|---|---|---|
| ADTypes | `1` | `1.18` |
| DiffEqBase | `6.36, 7` | `7.2` |
| DiffResults | `1.0` | `1.1` |
| Distributions | `0.25` | `0.25.123` |
| DocStringExtensions | `0.9` | `0.9.5` |
| DynamicHMC | `3` | `3.4.4` |
| ForwardDiff | `0.10, 1` | `1.0.1` |
| LabelledArrays | `1.0` | `1.19` |
| LogDensityProblemsAD | `1` | `1.0.1` |
| MacroTools | `0.5` | `0.5.16` |
| OrdinaryDiffEq | `6, 7` | `7` |
| PDMats | `0.11` | `0.11.35` |
| ParameterizedFunctions | `5` | `5.24` |
| Reexport | `1.0` | `1.2.2` |
| Requires | `1.0` | `1.3` |
| SciMLStructures | `1` | `1.7` |
| StanSample | `6, 7` | `7` |
| StatsBase | `0.34` | `0.34.6` |
| SteadyStateDiffEq | `1.9, 2` | `2.10.1` |
| TransformVariables | `0.8` | `0.8.15` |
| TransformedLogDensities | `1` | `1.0.3` |

## Verification (local, Julia 1.10 lts)

- **Downgrade-pinned merged resolve** (the failing CI check) now SUCCEEDS.
- **Minimality**: lowering any single floor reintroduces the `Unsatisfiable` error (each raise was verified to be the lowest workable bound).
- **Normal (non-downgrade) resolve** still SUCCEEDS on both Julia 1.10 and 1.12, selecting the latest releases (ForwardDiff 1.4.1, Distributions 0.25.127, ModelingToolkit 11.28.0, OrdinaryDiffEq 7.0.1, Turing 0.45.0, etc.) — so the regular test matrix is unaffected.

Please ignore until reviewed by @ChrisRackauckas


ChrisRackauckas-Claude:
## Status: floors are correct, but this PR does not (and cannot) flip the Downgrade check green

Followed up on why **Downgrade / Downgrade Tests - Core** is still red on this branch. The floor raises here are right — the merged downgrade env resolves cleanly when targeted at Julia 1.11/1.12. The residual red is a **separate, upstream-only blocker**, not a missing floor:

- The single unsatisfiable requirement is **`Turing`**. All `Turing ≥ 0.43.3` (incl. our floor `0.45.0`) and its modern deps (`DynamicPPL`, `Bijectors`, …) declare `[compat] julia = "1.10.8 - 1"`.
- The centralized strict downgrade collapses `julia-version: lts` to the **minor** `1.10` and runs `Resolver.jl --min=@deps --julia=1.10`, which spans the whole `1.10.x` line. With the floors pinned, that resolution **succeeds at `--julia=1.11`/`1.12` and fails only at `--julia=1.10`** (reproduced locally with the real Resolver against the live registry; a minimal `Turing="0.45"` + `julia="1.10"` project reproduces it in isolation).
- Setting `[compat] julia = "1.10.8"` here does not help — Resolver's Julia-candidate branch ignores `project_compat` and the action reduces the floor to minor `1.10` anyway.

So no `[compat]` edit in this repo turns the strict `lts` check green; the fix must land in `julia-actions/julia-downgrade-compat` / `Resolver.jl` (resolve at the runtime patch / honor a patch-level Julia floor) or via the SciML LTS alias advancing past 1.10. Full analysis posted on #370.

The floor raises in this PR remain valid and worth keeping (they fix genuine stale lower bounds and let the merged env resolve at 1.11/1.12). Please ignore until reviewed by @ChrisRackauckas.

Issue #391, opened 2026-06-25
QA: make DiffEqBayes' 68 implicit imports explicit (ExplicitImports no_implicit_imports)


ChrisRackauckas-Claude:
Tracking issue for the `no_implicit_imports` ExplicitImports check, currently marked `@test_broken` (via `ei_broken = (:no_implicit_imports,)`) in the QA group.

`DiffEqBayes` pulls **68 names** implicitly through heavy `using BigDep` lines in `src/DiffEqBayes.jl` (`using DiffEqBase, Distributions, Turing, MacroTools, RecursiveArrayTools, ModelingToolkit, ..., DynamicHMC, TransformVariables, ...`). Making them explicit is a large refactor with real risk (e.g. `Turing` re-exports many `Distributions` names; `using DiffEqBase` surfaces `SciMLBase`), so it is deferred rather than done as a mass rewrite in the QA-conversion PR.

The names ExplicitImports would have us import explicitly (from `check_no_implicit_imports`):

```
using DiffEqBase: DiffEqBase, EnsembleProblem, SciMLBase, remake
using Distances: Distances
using DocStringExtensions: DocStringExtensions, FIELDS, SIGNATURES, TYPEDEF
using DynamicHMC: DynamicHMC, mcmc_with_warmup
using LinearAlgebra: LinearAlgebra, Diagonal
using LogDensityProblemsAD: LogDensityProblemsAD
using MacroTools: MacroTools
using ModelingToolkit: ModelingToolkit, parameters, solve
using Optim: Optim
using Parameters: Parameters, @unpack
using Random: Random
using RecursiveArrayTools: RecursiveArrayTools
using Requires: Requires
using SciMLStructures: SciMLStructures
using StanSample: StanSample, SampleModel, read_samples, stan_sample
using TransformVariables: TransformVariables, as, asℝ₊
using TransformedLogDensities: TransformedLogDensities, TransformedLogDensity
using Turing: Turing, Bernoulli, Beta, BetaBinomial, Binomial, Categorical,
              Cauchy, Chisq, Distributions, Exponential, Frechet, Gamma,
              GeneralizedPareto, Gumbel, Hypergeometric, InverseGamma, Laplace,
              LogNormal, MCMCSerial, MvNormal, NamedDist, NegativeBinomial,
              Normal, Pareto, Poisson, Rayleigh, TDist, Truncated, Uniform,
              VonMises, Weibull, logpdf, params, sample
```

[...]

PR #400, opened 2026-07-22
Fix issue #391: explicit imports for DiffEqBayes

ArpanC6:
This PR fixes issue #391.

**What I changed**

In src/DiffEqBayes.jl I replaced using BigDep with explicit imports (using BigDep: name1, name2) so that only the names used in the file are imported. This avoids implicit imports and makes the code clearer.

In test/qa/qa.jl, I removed :no_implicit_imports from the broken tests list since this check now passes.

**Testing**

I tested the changes by running check_no_implicit_imports(DiffEqBayes) which now passes.

I also ran the full test suite with ] test. Turing tests all passed (14/14). DynamicHMC had 8 pass and 2 broken, which were already broken before my changes.

This issue was marked to wait for Chris Rackauckas's review. I followed the suggested approach and the changes work as expected.


ArpanC6:
@ChrisRackauckas  I've opened PR #400 for the explicit imports refactor (#391). check_no_implicit_imports passes and all tests are green. Ready for review when you have time.

ChrisRackauckas (approved):

ArpanC6:
Fixed the QA failure I had imported too many names from Turing that weren't actually used in the codebase. Removed the stale ones. The Core (Julia pre) failure looks like a pre existing numerical/stochastic issue in the Turing tests unrelated to this PR. Let me know if anything else needs adjustment.

PR #408, opened 2026-08-15
docs: document public Bayesian inference API

ChrisRackauckas-Claude:
Please ignore this PR until it has been reviewed by @ChrisRackauckas.

## What changed

- Export and render the user-facing `StanODEData` marker and `StanResult` return type.
- Expand the three exported inference docstrings with SciMLStyle sections for arguments, keyword arguments, returns, errors, fields, and usage examples.
- Remove `warnonly` and disabled doctests from the Documenter configuration.
- Keep Stan examples as rendered usage snippets because they require an external CmdStan installation; strict docs checks remain enabled.
- Render `DynamicHMCPosterior` and the module docstring so strict CheckDocument has no missing public entries.

## Verification

- `timeout 3600 env GROUP=QA julia --startup-file=no --project=. -e 'using Pkg; Pkg.test()'`: `QA/qa.jl | 21 | 21 | 3m59.4s`.\n- `julia --startup-file=no --project=docs -e 'using Pkg; Pkg.develop(path=pwd()); include("docs/make.jl")'`: passed doctests, cross-references, CheckDocument, and HTML rendering.\n- `typos` over all changed source and documentation files: passed.\n- `git diff --check`: passed.\n- Runic format comparison with the installed Runic formatter: passed; installed Runic 1.0.0 has no `check` subcommand.\n\n## Known unrelated gap\n\nThe targeted Core group still has a pre-existing `UndefVarError` for `SteadyStateProblem` at `test/turing.jl:120`; that investigation is being handled separately and is not included here.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -11,9 +11,7 @@ makedocs(
     sitename = "DiffEqBayes.jl",
     authors = "Chris Rackauckas, Vaibhav Kumar Dixit et al.",
     clean = true,
-    doctest = false,
     modules = [DiffEqBayes],
-    warnonly = true,
     format = Documenter.HTML(
         assets = ["assets/favicon.ico"],
         canonical = "https://docs.sciml.ai/DiffEqBayes/stable/"
--- a/docs/pages.jl
+++ b/docs/pages.jl
@@ -1,5 +1,6 @@
 pages = [
     "index.md",
+    "Interfaces" => "interfaces.md",
     "Methods" => "methods.md",
     "Examples" => ["examples.md", "examples/pendulum.md"],
 ]
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -37,7 +37,7 @@ data = convert(Array, randomized)
 
 ### Stan
 
-```@example all
+```julia
 using StanSample #required for using the Stan backend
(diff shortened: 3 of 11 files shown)


ChrisRackauckas-Claude:
Follow-up audit pushed in commit 5bcba15.

- Added on-definition SciMLStyle docs/examples for exported StanODEData and StanResult, expanded the three inference API docstrings, and added rendered API/interface pages.
- Added a documented developer-interface contract for internal DynamicHMCPosterior; it is explicitly not user-facing.
- Removed the old warnonly/disabled-doctest configuration. Strict local Documenter build passes through doctests, cross-references, missing-doc checks, and HTML rendering.
- Added the generic public-interface Core test. After correcting the chain assertion to use size(chain, 1), it passes locally.
- Local Runic, typos, and git diff --check pass; local QA was 21/21.
- #408 Core on its own branch reaches 13/14 and fails only at test/turing.jl:120 because SteadyStateProblem is not imported. That is exactly the independent #409 fix.
- #409 was verified independently against clean master: local Core 14/14 and the hosted full matrix passed.

Merge order: merge #409 first, then rebase/update #408 and merge #408. The source PR changes contain documentation/API declaration work; the only non-documentation PR-local addition here is the generic contract test.

